PR #1, opened 2018-01-30
Add FEC functions

qiaozhigang:


p2p/fechelper.go
```diff
@@ -0,0 +1,2133 @@
...
+)
+
+//CFECHelper helper class for fec
+type CFECHelper struct {
+	canVec         [255]*VBitVec
+	bitLen, fecLen uint
+}
+
+//Init initialize helper.
+func (h *CFECHelper) Init(_bitLen uint, _fecLen uint) bool {
```
WTer: This function is so long, and more duplicated logic, suggest to refine it.

qiaozhigang: we will fix in future.

PR #6, opened 2018-02-02
Add PoW engine

WTer:
Interface and implementation.


consensus/pow/sha256worker.go
```diff
@@ -0,0 +1,64 @@
...
+    "crypto/sha256"
+    "math"
+    "math/big"
+)
+
+// data isathe pointer to the target block data,
+// target is the goal, which means we need to mine a result and its hash must be less than the target.
+type SHA256Worker struct {
+    Data   []byte
+    Nonce  int
```
lucusfly: nonce may be bigger than int.Max

consensus/pow/sha256worker.go
```diff
@@ -0,0 +1,64 @@
...
+    return data
+}
+
+// Loop nonce to find the target value that meet the requirement.
+func (worker *SHA256Worker) Producer() int {
+    nonce := 0
+    var hash [32]byte
+    var hashInt big.Int
+
+    for nonce < math.MaxInt64 {
```
lucusfly: math.MaxInt64?

consensus/pow/utils.go
```diff
@@ -0,0 +1,18 @@
+package consensus
+
```
lucusfly: move to common folder

consensus/pow/utils.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+    "bytes"
+    "encoding/binary"
+    "log"
+)
+
+// Converts a int64 value into a byte array
+func Int2Hex(num int64) []byte {
```
lucusfly: Int2Hex should be Int2Bytes

consensus/pow/worker.go
```diff
@@ -0,0 +1,10 @@
+package consensus
+
+// Worker is an PoW engine.
+type Worker interface {
+    // Returns the current mining result of a PoW consensus engine.
+    Producer() int
```
lucusfly: Producer() int
better to return a string if you want it to be general

consensus/pow/worker.go
```diff
@@ -0,0 +1,10 @@
+package consensus
+
```
lucusfly: move up to consensus folder for interface

consensus/pow/sha256worker.go
WTer: Good catch, fixed it.

consensus/pow/sha256worker.go
WTer: I changed the definition of nonce from int to int64.

consensus/pow/utils.go
WTer: consensus is a independent component, there is also consensus folder in ETH. so I would prefer to keep it.

consensus/pow/utils.go
WTer: nice.

consensus/pow/worker.go
WTer: nice suggestion. fixed.

consensus/pow/worker.go
WTer: just keep it.

consensus/pow/sha256worker.go
```diff
@@ -0,0 +1,64 @@
...
+    "bytes"
+    "crypto/sha256"
+    "math"
+    "math/big"
+)
+
+// data isathe pointer to the target block data,
+// target is the goal, which means we need to mine a result and its hash must be less than the target.
+type SHA256Worker struct {
+    Data   []byte
```
boqiu: Comment is required for all exported type/variable, and better to follow the golint  guideline.

WTer: nice.

common/converter.go
```diff
@@ -0,0 +1,18 @@
+package common
+
+import (
+    "bytes"
+    "encoding/binary"
+    "log"
+)
+
+// Int2Bytes Converts a int64 value into a byte array
+func Int2Bytes(num int64) []byte {
```
boqiu: Confused func name, what about ConvertInt64ToBytes?

consensus/pow/sha256/sha256worker.go
```diff
@@ -0,0 +1,66 @@
+package sha256
```
boqiu: How about change file name to worker.go

common/converter.go
WTer: sure, changed.

consensus/pow/sha256/sha256worker.go
WTer: nice suggestion.

consensus/pow/worker.go
```diff
@@ -0,0 +1,10 @@
+package pow
+
+// Worker is an PoW engine.
+type Worker interface {
+    // Returns the current mining result of a PoW consensus engine.
+    Producer() string
+
+    // Verify whether the mining result is meet the requirement.
+    Validator() bool
```
boqiu: Validate

consensus/pow/worker.go
```diff
@@ -0,0 +1,10 @@
+package pow
+
+// Worker is an PoW engine.
+type Worker interface {
+    // Returns the current mining result of a PoW consensus engine.
+    Producer() string
+
+    // Verify whether the mining result is meet the requirement.
```
boqiu: redundant "is"

consensus/pow/worker.go
WTer: changed.

consensus/pow/worker.go
WTer: removed.

PR #7, opened 2018-02-02
Addlog

lucusfly:


log/log.go
```diff
@@ -0,0 +1,78 @@
...
+    "fmt"
+    "github.com/sirupsen/logrus"
+    "os"
+    "os/user"
+    "path/filepath"
+)
+
+var log *logrus.Logger
+
+// PanicLevel level, highest level of severity. Logs and then calls panic with the
```
WTer: nice to add a space between Panic and Level:
PanicLevel  --> Panic Level

to meet golint format requirement.

log/log.go
```diff
@@ -0,0 +1,78 @@
+/**
+*  @file
+*  @copyright defined in go-seele/LICENSE
+ */
+package log
```
WTer: add unittests for this package.

log/log.go
```diff
@@ -0,0 +1,78 @@
...
+    "os"
+    "os/user"
+    "path/filepath"
+)
+
+var log *logrus.Logger
+
+// PanicLevel level, highest level of severity. Logs and then calls panic with the
+// message passed to Debug, Info, ...
+func Panic(args ...interface{}) {
```
WTer: all these functions are thread-safe?

log/log.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// DebugLevel level. Usually only enabled when debugging. Very verbose logging.
+func Debug(args ...interface{}) {
+    log.Debugln(args)
+}
+
+func init()  {
+    usr, err := user.Current()
+    if err != nil {
```
WTer: why do we get current user and set it as the required condition for log initializaiton?

log/log.go
```diff
@@ -0,0 +1,78 @@
...
+    seeleFolder := filepath.Join(usr.HomeDir, "seelelog")
+    if _, err := os.Stat(seeleFolder); err != nil {
+        if err = os.Mkdir(seeleFolder, os.ModeDir); err != nil {
+            fmt.Println("create log folder failed. ", err)
+            return
+        }
+    }
+
+    seeleLog := filepath.Join(seeleFolder, "log.txt")
+    file, err := os.OpenFile(seeleLog, os.O_CREATE | os.O_WRONLY, 0666)
```
WTer: what's happened when this function called multiple times by the same user?

log/log.go
lucusfly: add log_test

log/log.go
lucusfly: sure

log/log.go
lucusfly: yes, logrus are thread-safe

log/log.go
lucusfly: we need find a log file, by default I use $USER/seelelog/log.txt as log file, maybe we could have a better way to find it later

log/log.go
lucusfly: the init function is called automatically when package initialized. no one should call it.

PR #10, opened 2018-02-03
Add and refine unittests for VPacket

WTer:


p2p/vpacket.go
```diff
@@ -1,17 +1,19 @@
...
  */
 package p2p
 
-import "encoding/binary"
+import  (
+    "encoding/binary"
+)
 
 const (
     //VPacketHeadLen head length in net-package
```
herove: Add a space at the beginning, pay attention to others

WTer: nice suggestion. fixed it.

qiaozhigang: dataLen don't need send to peer, because in the receiver end it can be calculated by udp package size and VPacketHeadLen.

WTer: sure, I already reverted to original version.

PR #13, opened 2018-02-04
add Travis check status

lucusfly:


.travis.yml
```diff
@@ -0,0 +1,2 @@
+language: go
+script: go test ./...
```
WTer: add new line at end of this file.

log/log.go
```diff
@@ -49,7 +49,7 @@ func Debug(args ...interface{}) {
 }
 
 func init()  {
-    usr, err := user.Current()
+   /* usr, err := user.Current()
```
WTer: why do you comment these statements?

log/log.go
```diff
@@ -69,10 +69,11 @@ func init()  {
...
-
+*/
     logrus.SetFormatter(&logrus.TextFormatter{})
 
     log = logrus.New()
-    log.Out = file
+    //log.Out = file //use std out for temp
+    log.Out = os.Stdout
     log.SetLevel(logrus.DebugLevel)
 }
```
WTer: add new line at the end of this file.

.travis.yml
lucusfly: added

log/log.go
lucusfly: travis will fail whit test when create file, just comments them for quick pass

log/log.go
lucusfly: added

.travis.yml
lucusfly: added

log/log.go
lucusfly: added

log/log.go
lucusfly: travis will fail whit test when create file, just comments them for quick pass

PR #14, opened 2018-02-04
Add unittests for SpeedMeter

WTer:


p2p/speedmeter_test.go
```diff
@@ -10,10 +10,48 @@ import (
...
-    sp.Feed(128 * 1024)
-    time.Sleep(300)
-    sp.Feed(128 * 1024)
-    fmt.Println("speedMeter:", sp.GetRate())
+func Test_NewSpeedMeter(t *testing.T) {
+    var step uint64 = 5
+    var itemsNum uint = 10
+    v1 := NewSpeedMeter(step, itemsNum)
+
+    if v1.step != step || v1.itemsNum != itemsNum {
```
lucusfly: use check.Assert?

p2p/speedmeter_test.go
```diff
@@ -10,10 +10,48 @@ import (
...
+    }
+
+    rate := v1.GetRate()
+    if rate != 0 {
+        fmt.Println("NewSpeedMeter rate:", rate)
+        t.Fail()
+    }
+
+    var num uint = 10
+    time.Sleep(1000 * time.Millisecond)
```
lucusfly: why you need sleep?

p2p/speedmeter_test.go
```diff
@@ -10,10 +10,48 @@ import (
...
+    if v1.preFeedTick <= 0 {
+        fmt.Println("NewSpeedMeter preFeedTick:", v1.preFeedTick)
+        t.Fail()
+    }
+
+    for _, ele := range v1.itemArr {
+        if ele.tick == 0 {
+            fmt.Println("NewSpeedMeter Feed:", ele)
+            t.Fail();
+        }
```
lucusfly: use check.Assert only need one line for 52-55 lines

p2p/speedmeter_test.go
WTer: the following statement is Feed(), in which the time.Now will be called, so I use this simple way to simulate it.

p2p/speedmeter_test.go
WTer: Currently I use testing.T.  we can refine it to check later.

PR #18, opened 2018-02-06
Add common and crypto function

lucusfly:


common/hash.go
```diff
@@ -0,0 +1,27 @@
...
+	var a Hash
+	a.SetBytes(b)
+	return a
+}
+
+func StringToAddress(s string) Hash {
+	return BytesToAddress([]byte(s))
+}
+
+// Sets the hash to the value of b. If b is larger than len(a) it will panic
```
WTer: comments should be started with SetBytes

common/hash.go
```diff
@@ -0,0 +1,27 @@
+/**
```
WTer: nice to add unittests for these files.

common/hash.go
lucusfly: resolved

common/hash.go
lucusfly: add test

PR #21, opened 2018-02-06
remove check package and use another assert package

lucusfly:


common/rlpwrapper_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/rlp"
-	"gopkg.in/check.v1"
+	"github.com/magiconair/properties/assert"
```
WTer: that's cool.

PR #23, opened 2018-02-08
add build check and rename method

lucusfly:


common/rlpwrapper_test.go
```diff
@@ -54,15 +54,15 @@ func Test_RLP(t *testing.T) {
 
 // test gob effective
 func Test_Encoding(t *testing.T) {
-	data, err := Encoding(&s)
+	data, err := Serialize(&s)
 	if err != nil {
 		fmt.Println(err)
```
WTer: add Assert.Error

common/rlpwrapper_test.go
```diff
@@ -54,15 +54,15 @@ func Test_RLP(t *testing.T) {
...
 	}
 
 	fmt.Println(len(data))
 
 	nst := Student{}
-	err = Decoding(data, &nst)
+	err = Deserialize(data, &nst)
 
 	fmt.Printf("%v\n", nst)
```
WTer: using Assert instead of printf.

PR #28, opened 2018-02-10
add vendor - google uuid

boqiu:


vendor/vendor.json
```diff
@@ -39,6 +39,12 @@
 			"revision": "c2828203cd70a50dcccfb2761f8b1f8ceef9a8e9",
 			"revisionTime": "2018-01-10T05:33:47Z"
 		},
+		{
```
WTer: Why did you add this section?

boqiu: This is auto-generated by govendor which a standard tool to import packages into project.

WTer: got it, but its revisionTime is a little strange.

PR #33, opened 2018-02-12
init contract - reg/invoke

boqiu:
Smart contract component startup. Implement reg/invoke with mock service.

### DIFF
--- a/contract/external.go
+++ b/contract/external.go
@@ -1,0 +1,60 @@
+/**
+* This a temp file that mock the external APIs used in smart contract component.
+* It will be removed when blockchain infrastructure constructed, including account,
+* security, RPC, consensus and blockchain storage components.
+*
+* @file
+* @copyright defined in go-seele/LICENSE
+ */
+package contract
+
+import (
+	"github.com/seeleteam/go-seele/crypto"
+)
+
+const (
+	txTypeContractReg int = iota
+	txTypeContractCall
+	txTypeContractDel
+)
+
+// Account wraps the information of smart contract account.
+type Account struct {
+	AccountAddress string
+	CodeAddress    string
+	State          []byte
+}
+
+// Transaction wraps the smart contract transaction.
+type Transaction struct {
+	from    string
+	to      string
+	txType  int
+	payload []byte
+	sig     *crypto.Signature
+}
+
+// TransactionService is the interface for all transaction related operations.
+type TransactionService interface {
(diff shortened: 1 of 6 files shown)


contract/external.go
```diff
@@ -0,0 +1,60 @@
...
+* @copyright defined in go-seele/LICENSE
+ */
+package contract
+
+import (
+	"github.com/seeleteam/go-seele/crypto"
+)
+
+const (
+	txTypeContractReg int = iota
```
WTer: what's iota?

boqiu: auto increase value start with 0, this is golang syntax for enum definition.

contract/external.go
```diff
@@ -0,0 +1,60 @@
...
+	txTypeContractCall
+	txTypeContractDel
+)
+
+// Account wraps the information of smart contract account.
+type Account struct {
+	AccountAddress string
+	CodeAddress    string
+	State          []byte
```
WTer: add some comments for these members.

WTer: I suppose State's type is something like enum

contract/external.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Transaction wraps the smart contract transaction.
+type Transaction struct {
+	from    string
+	to      string
+	txType  int
+	payload []byte
+	sig     *crypto.Signature
```
WTer: seems rename sig as sign should be better and common

contract/external.go
boqiu: All the code in external.go are temp, so please ignore the comment.

contract/external.go
boqiu: All the code in external.go are temp, so please ignore the comment.

contract/handler.go
```diff
@@ -0,0 +1,50 @@
...
+		log.Error("Transaction signature is invalid, and will not be archived into blockchain.")
+		return
+	}
+
+	switch tx.txType {
+	case txTypeContractReg:
+		onRegisterContract(tx)
+	case txTypeContractCall:
+		onInvokeContract(tx)
+	case txTypeContractDel:
```
WTer: add default case

contract/external.go
boqiu: The state should be of type object, but use []byte for serialization.

contract/handler.go
boqiu: unnecessary, do not handle not supported tx type.

PR #36, opened 2018-02-27
add makefile

lucusfly:
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/36)


Makefile
```diff
@@ -0,0 +1,3 @@
+discovery:
+	go build .\cmd\discovery
+	@echo "Done discovery building"
```
WTer: take care this warn.

PR #38, opened 2018-02-28
Add 2 functions Has and Delete to leveldb

WTer:
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/38)

### DIFF
--- a/database/leveldb/leveldb.go
+++ b/database/leveldb/leveldb.go
@@ -48,3 +48,19 @@ func (db *levelDB) Put(key []byte, value []byte) error {
 func (db *levelDB) PutString(key string, value string) error {
 	return db.Put([]byte(key), []byte(value))
 }
+
+func (db *levelDB) Has(key []byte) (ret bool, err error) {
+	return db.db.Has(key, nil)
+}
+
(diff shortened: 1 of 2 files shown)


herove:
Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[database/leveldb/leveldb.go, line 57 at r2](https://reviewable.io:443/reviews/seeleteam/go-seele/38#-L6PG9F8zR1S71v--l0X:-L6PG9F8zR1S71v--l0Y:biewc90) ([raw file](https://github.com/seeleteam/go-seele/blob/993c5875a5b5f3cf019c919a26e56a9fcb257759/database/leveldb/leveldb.go#L57)):*
> ```Go
> 
> func (db *levelDB) HasString(key string) (ret bool, err error) {
> 	return db.db.Has([]byte(key), nil)
> ```


In the function "put", the first db is a wrapper, in function "putstring" calls "put", for format, it is recommended to modify db.db to db.has(current class). Never mind.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/38)*

PR #62, opened 2018-03-06
fix make all doesn't work

chenhuaying:
update Makefile for make all

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/62)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,10 @@
+all: discovery node
 discovery:
 	go build ./cmd/discovery
 	@echo "Done discovery building"
+
+node:
+	go build -o seele-node ./cmd/node 
+	@echo "Done node building"
+
+.PHONY: discovery node


Makefile
```diff
@@ -1,3 +1,10 @@
+all: discovery node
 discovery:
 	go build ./cmd/discovery
 	@echo "Done discovery building"
+
+node:
+	go build -o seele-node ./cmd/node
```
lucusfly: please don't specific output file. For different os, the file extension is different

chenhuaying: oh，I forgot other OS like windows. I modify the output file name because the default is conflicted with the directory "node" in the project root dir. May be we can check OS or add a bin directory and put the compiled files in this dir?

WTer:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/62)*

Makefile
lucusfly: Anyway, I think we could only support linux env

PR #85, opened 2018-03-19
add trie;implement MPT

jessonzheng:
add trie for state db;It is a merkle patricia tree using by state db & storage db(contract account)

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/85)


herove:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[trie/node.go, line 1 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xItKwAzHc7ybqKrbS:-L7xItKwAzHc7ybqKrbT:bxabg1d) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/node.go#L1)):*
> ```Go
> package trie
> ```

/**
*  @file
*  @copyright defined in go-seele/LICENSE
 */

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

gongliaoan:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[trie/trie.go, line 29 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xSmbYOFTfvQ3-60nY:-L7xSmbYOFTfvQ3-60nZ:b-hqmw4j) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L29)):*
> ```Go
> // NewTrie new a trie tree
> func NewTrie(root common.Hash, prefix []byte, db database.Database) (*Trie, error) {
> ```

blank line should delete

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

gongliaoan:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[trie/trie.go, line 89 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xTUgomE3wyg5hybrG:-L7xTUgomE3wyg5hybrH:b38u654) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L89)):*
> ```Go
> 
> // Hash return the hash of trie
> func (t Trie) Hash() common.Hash {
> ```

is t *Trie?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

gongliaoan:
Review status: 0 of 3 files reviewed at latest revision, 5 unresolved discussions.

---

*[trie/trie.go, line 20 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xXO85Ui5Gx6g_Ceav:-L7xXO85Ui5Gx6g_Ceaw:b-gn0ntf) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L20)):*
> ```Go
> )
> 
> //Trie is a Merkle Patricia Trie
> ```

comment start with a blank.
// Trie ...

---

*[trie/trie.go, line 397 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xavVMFHgLc5AZV1Bc:-L7xavVMFHgLc5AZV1Bd:b-aqq8fr) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L397)):*
> ```Go
> 	case 2:
> 		return t.decodeLeafNode(hash, vals)
> 	case 3:
> ```

is there only 3 values?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[trie/node.go, line 5 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xga1gptLthu7Ah8BG:-L7xga1gptLthu7Ah8BH:b-557lda) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/node.go#L5)):*
> 17

nice to add comments to explain why it's 17 instead of other values, for performance, or other purposes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[trie/trie.go, line 51 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xhqGB8VShdk2oMZI0:-L7xhqGB8VShdk2oMZI1:b-weqepr) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L51)):*
> if err == nil

what's happened when err != nil? do we need handle such error case?

---

*[trie/trie.go, line 54 at r1](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L7xhlpth4l0QNpuodX_:-L7xhlpth4l0QNpuodXa:b2pix5b) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L54)):*
>  

remove this blank line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

boqiu:
I recommend to use polymorphism to avoid so many "type switch-case" statements. Long method is hard to read/maintain.

---

Review status: 0 of 3 files reviewed at latest revision, 8 unresolved discussions.

---

*[trie/node.go, line 5 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L7xga1gptLthu7Ah8BG:-L80IUg01V558qOC-RQj:b-fadrev) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/node.go#L5)):*
<details><summary><i>Previously, jessonzheng wrote…</i></summary><blockquote>

for 0-f branches + value node; reduce the high of tree for performance
</blockquote></details>

1) Unnecessary to export this const variable.
2) Suggest to use a meaningful name, e.g. numBranchNodes.

---

*[trie/node.go, line 12 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80JF9D80TpUhuXB1PM:-L80JF9D80TpUhuXB1PN:b-4hfb9h) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L12)):*
> ```Go
> 	NumberChildren int = 17 // for 0-f branches + value node; reduce the high of tree for performance
> 	// LengthOfNodeHash length of node hash
> 	LengthOfNodeHash int = 32
> ```

If node hash type is common.Hash, then suggest to use the existing const variable common.HashLength that defined in common/hash.go

---

*[trie/node.go, line 16 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80J_6c8U8JVUovWWe9:-L80J_6d051AuZMZKonb:b-lyz7hs) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L16)):*
> ```Go
> 
> // Noder interface for node
> type Noder interface {
> ```

Is it necessary to export this interface? Please double check all the defined structs/interfaces in this PR. Ideally, MPT client should not be aware of these internal data structures.

---

*[trie/node.go, line 17 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80K5_n9zQ-FxBog7Hb:-L80K5_n9zQ-FxBog7Hc:b-f7s4w0) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L17)):*
> []byte

common.Hash? Just double confirm, please ignore it if you plan to avoid unnecessary hash-value-copy.

---

*[trie/node.go, line 18 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Kkgs0rC-A0ic_B8L:-L80Kkgs0rC-A0ic_B8M:b96c8u2) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L18)):*
> ```Go
> type Noder interface {
> 	Hash() []byte
> 	IsDirty() bool
> ```

Suggest to add necessary comments to describe what is dirty mean and when dirty node occurs?

---

*[trie/node.go, line 27 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80LrBm6Nyrs46HLgPA:-L80LrBm6Nyrs46HLgPB:b5h0apf) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L27)):*
> ```Go
> }
> 
> // ExtendNode is extend node struct.for root,extend node
> ```

necessary whitespaces missed.

---

*[trie/node.go, line 30 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80LeDs5XlcoZCG-vQJ:-L80LeDs5XlcoZCG-vQK:bhdnwkz) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L30)):*
> ```Go
> type ExtendNode struct {
> 	Node
> 	Key      []byte // for shared nibbles or key-end
> ```

it's shared nibbles for extend node, not key-end

---

*[trie/node.go, line 37 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Lwap8lyapban418I:-L80Lwap8lyapban418J:bh8v9tf) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L37)):*
> ```Go
> type LeafNode struct {
> 	Node
> 	Key   []byte // for shared nibbles or key-end
> ```

it's key-end for leaf node, not shared nibbles.

---

*[trie/node.go, line 44 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80MTt1EyWdWzF41aYD:-L80MTt1EyWdWzF41aYE:b-1nlybt) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L44)):*
> ```Go
> type BranchNode struct {
> 	Node
> 	Children [NumberChildren]Noder
> ```

This last element is for value, not Noder type, right?

---

*[trie/node.go, line 48 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80NIefAqCmlu3CXW8p:-L80NIefAqCmlu3CXW8q:broxscm) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L48)):*
> datbase

typo: database

---

*[trie/trie.go, line 397 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L7xavVMFHgLc5AZV1Bc:-L80VxFJ5PECgz1T8Rsu:bs91hlz) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L397)):*
<details><summary><i>Previously, jessonzheng wrote…</i></summary><blockquote>

Yes,the number means how many numbers in the struct that encoded by rlp
</blockquote></details>

Please use const variable instead of such magic number.

---

*[trie/trie.go, line 26 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Th-5EQPBidv8jnjp:-L80Th-5EQPBidv8jnjq:b23x3b5) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L26)):*
> ```Go
> 
> // Trie is a Merkle Patricia Trie
> type Trie struct {
> ```

Do you consider to ensure the thread-safe operations?

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L80OKEcAc0VVnne9wPP:b-2pzfmn) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
> ```Go
> 
> // NewTrie new a trie tree
> func NewTrie(root common.Hash, prefix []byte, db database.Database) (*Trie, error) {
> ```

What is prefix mean? Any suggested value to create a MPT?

---

*[trie/trie.go, line 39 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80O_dWAHA3FgYhaZ8r:-L80O_dWAHA3FgYhaZ8s:b-w7vbng) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L39)):*
> common.Hash{}

common.EmptyHash

---

*[trie/trie.go, line 40 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80P5uJ8LDcxBmeZUtR:-L80P5uJ8LDcxBmeZUtS:b75mu7r) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L40)):*
> root[:]

better to use existing API to convert Hash to []byte

---

*[trie/trie.go, line 47 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Qw544q2Ff-pFw01W:-L80Qw544q2Ff-pFw01X:b-3ah1zj) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L47)):*
> ```Go
> 	}
> 
> 	return trie, nil
> ```

In which case will create a tree with null root? It's really confused.

If possible, the root node should never be null to simplify the code.

---

*[trie/trie.go, line 60 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RR5DAEzFurGnazHE:-L80RR5DAEzFurGnazHF:b-9az0iy) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L60)):*
> ```Go
> }
> 
> // Delete delete node with key in the trie
> ```

Suggest to add necessary comment about the return value.

---

*[trie/trie.go, line 74 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RzH3EcjGsmr2Lb0T:-L80RzH3EcjGsmr2Lb0U:b-r1kz86) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L74)):*
> ```Go
> 
> // Get get the value by key
> func (t *Trie) Get(key []byte) []byte {
> ```

maybe return ([]byte, bool) is better for client to check if the result is valid.

---

*[trie/trie.go, line 78 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RpPc3JxBvaPPL7_U:-L80RpPc3JxBvaPPL7_V:bnnmct) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L78)):*
> val != nil &&

len(val) is enough, len(nil) is 0

---

*[trie/trie.go, line 99 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80SaxZ6mqVByGbWp_A:-L80SaxZ6mqVByGbWp_B:b-w7vbng) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L99)):*
> common.Hash{}

common.EmptyHash

---

*[trie/trie.go, line 103 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80SjivBtKi2l9jWV3a:-L80SjivBtKi2l9jWV3b:b-gfstfe) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L103)):*
> ```Go
> 
> // Commit commit the dirty node to database
> func (t *Trie) Commit(batch database.Batch) (common.Hash, error) {
> ```

The code below is the same with previous API, copy mistake? If not, please avoid copy-paste.

---

*[trie/trie.go, line 113 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80vIrU1CZ2eh75DB16:-L80vIrU1CZ2eh75DB17:bw5k2pe) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L113)):*
> ```Go
> }
> 
> func (t *Trie) hash(node Noder, buf *bytes.Buffer, sha hash.Hash, batch database.Batch) []byte {
> ```

unnecessary buf & sha parameter, please refer to crypto.Keccak256Hash for hash calculation.

---

*[trie/trie.go, line 121 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80vmQMBQad4egdDgQb:-L80vmQMBQad4egdDgQc:b-txxc8j) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L121)):*
> ```Go
> 	}
> 	switch n := node.(type) {
> 	case *LeafNode:
> ```

Noder interface has Hash() API already, why not implement it in all kinds of Node? 

---

*[trie/trie.go, line 179 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80UDZU79lVDhxBWW3p:-L80UDZU79lVDhxBWW3q:b-h337kd) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L179)):*
> ```Go
> }
> 
> func (t *Trie) insert(node Noder, key []byte, value []byte) (bool, Noder, error) {
> ```

Suggest to add necessary comments for such long method.

---

*[trie/trie.go, line 181 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80UQBS4YXv9p-Bijpo:-L80UQBS4YXv9p-Bijpp:bph0kyt) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L181)):*
> ```Go
> func (t *Trie) insert(node Noder, key []byte, value []byte) (bool, Noder, error) {
> 	switch n := node.(type) {
> 	case *ExtendNode:
> ```

How about use polymorphsim to simplify the code, since such long method is really bad to read and hard to understand in a short time.

---

*[trie/trie.go, line 378 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80VdSnE_sRFEftiOyv:-L80VdSnE_sRFEftiOyw:b-w35plu) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L378)):*
> val == nil

len(val) == 0

---

*[trie/trie.go, line 393 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80VqmE6k0YTLHyCj_O:-L80VqmE6k0YTLHyCj_P:b-dpdr5v) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L393)):*
> ```Go
> 		return nil, err
> 	}
> 	switch n, _ := rlp.CountValues(vals); n {
> ```

error handling?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-:-L80x3gw9Ko7RiKjABC3:b-q4g5w2)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 33 unresolved discussions.

---

*[trie/trie_test.go, line 26 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80yD170g_4qr1ilfG-:-L80yD170g_4qr1ilfG0:bxo6i1f) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L26)):*
> ```Go
> 	db, err := leveldb.NewLevelDB(dir)
> 	if err != nil {
> 		panic(err)
> ```

delete the temp folder before panic.

---

*[trie/trie_test.go, line 52 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80zLrRD-LtwnVJ8r1G:-L80zLrRD-LtwnVJ8r1H:butskp9) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L52)):*
> ```Go
> 	value := trie.Get([]byte("12345678"))
> 	fmt.Println(string(value))
> 	assert.Equal(t, string(value), "testnew")
> ```

I think it's better to assert the trie structure (different nodes/values), instead of the Get API. Or add more tests about the data structure.

---

*[trie/trie_test.go, line 76 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80yyxxD8cEBjKkvBhK:-L80yyxxD8cEBjKkvBhL:bpr3zut) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L76)):*
> ```Go
> 	batch := db.NewBatch()
> 	trie.Commit(batch)
> 	batch.Commit()
> ```

Any assertion after commit? If no, it's unnecessary to commit the updates.

---

*[trie/trie_test.go, line 95 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81-0PN3W3-Sn4yuiDr:-L81-0PN3W3-Sn4yuiDs:b-8p9azy) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L95)):*
> ```Go
> 	trie.Update([]byte("24375879"), []byte("test5"))
> 	trie.Update([]byte("24375878"), []byte("test6"))
> 	trie.Update([]byte("24355879"), []byte("test7"))
> ```

Above Update statements are duplicated with previous test, suggest to define common test data.

And again, suggest to assert the data structure.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

jessonzheng:
I try it. But they are recursive functions. Node.go includes Trie  looks odd. I have split the long function.  

---

Review status: 0 of 3 files reviewed at latest revision, 37 unresolved discussions.

---

*[trie/node.go, line 5 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L7xga1gptLthu7Ah8BG:-L81f8l--iF_NL3rLWsi:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/node.go#L5)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

1) Unnecessary to export this const variable.
2) Suggest to use a meaningful name, e.g. numBranchNodes.
</blockquote></details>

Done.

---

*[trie/node.go, line 12 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80JF9D80TpUhuXB1PM:-L81fZNlDqlphtfMaPEr:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L12)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

If node hash type is common.Hash, then suggest to use the existing const variable common.HashLength that defined in common/hash.go
</blockquote></details>

Done.

---

*[trie/node.go, line 16 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80J_6c8U8JVUovWWe9:-L81gKnuBWdolP_-khOb:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L16)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Is it necessary to export this interface? Please double check all the defined structs/interfaces in this PR. Ideally, MPT client should not be aware of these internal data structures.
</blockquote></details>

Done.

---

*[trie/node.go, line 17 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80K5_n9zQ-FxBog7Hb:-L81dbLG89odmhCNBg1o:b-ohq8eg) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L17)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> []byte

common.Hash? Just double confirm, please ignore it if you plan to avoid unnecessary hash-value-copy.
</blockquote></details>

to avoid value copy,just be used by trie. 

---

*[trie/node.go, line 18 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Kkgs0rC-A0ic_B8L:-L81eMUX5r9btYBtNSLr:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L18)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Suggest to add necessary comments to describe what is dirty mean and when dirty node occurs?
</blockquote></details>

Done.

---

*[trie/node.go, line 27 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80LrBm6Nyrs46HLgPA:-L81dFgo3w54tdrrSxlM:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L27)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

necessary whitespaces missed.
</blockquote></details>

Done.

---

*[trie/node.go, line 30 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80LeDs5XlcoZCG-vQJ:-L81cf0g6RGJq6d2Gk_4:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L30)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

it's shared nibbles for extend node, not key-end
</blockquote></details>

Done.

---

*[trie/node.go, line 37 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Lwap8lyapban418I:-L81cekc7CYbf3duUkW9:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L37)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

it's key-end for leaf node, not shared nibbles.
</blockquote></details>

Done.

---

*[trie/node.go, line 44 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80MTt1EyWdWzF41aYD:-L81gPVe8BSJjnEINWVH:b-j1jp7x) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L44)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

This last element is for value, not Noder type, right?
</blockquote></details>

no.it is noder type

---

*[trie/node.go, line 48 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80NIefAqCmlu3CXW8p:-L81cJPPENq3vlBdF5SR:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/node.go#L48)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> datbase

typo: database
</blockquote></details>

Done.

---

*[trie/trie.go, line 397 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L7xavVMFHgLc5AZV1Bc:-L81ibcA-hc2BlAIFu1d:bayjr98) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/trie.go#L397)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Please use const variable instead of such magic number.
</blockquote></details>

the const variable is also odd.It is not  necessary.

---

*[trie/trie.go, line 26 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Th-5EQPBidv8jnjp:-L81W4hH028AY5TFRu-M:b-g6kj9m) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L26)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Do you consider to ensure the thread-safe operations?
</blockquote></details>

no, this should be considered by up level.

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L81bQoZA1etxSGUKG-g:b-b25i52) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

What is prefix mean? Any suggested value to create a MPT?
</blockquote></details>

prefix is used by different tree;such as account state tree,  contract state tree. they all use 32 bytes hash as key.  

---

*[trie/trie.go, line 39 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80O_dWAHA3FgYhaZ8r:-L81ax6XBQnLexS9_8nv:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L39)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> common.Hash{}

common.EmptyHash
</blockquote></details>

Done.

---

*[trie/trie.go, line 40 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80P5uJ8LDcxBmeZUtR:-L81bNxeDOi7OohoAHkf:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L40)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> root[:]

better to use existing API to convert Hash to []byte
</blockquote></details>

Done.

---

*[trie/trie.go, line 47 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80Qw544q2Ff-pFw01W:-L81alaC0t2g_7mdtXNF:b-t7u5z3) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L47)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

In which case will create a tree with null root? It's really confused.

If possible, the root node should never be null to simplify the code.
</blockquote></details>

the tree maybe not load because of error hash key.

---

*[trie/trie.go, line 60 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RR5DAEzFurGnazHE:-L81a_XTBO8SGUdBza7g:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L60)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Suggest to add necessary comment about the return value.
</blockquote></details>

Done.

---

*[trie/trie.go, line 74 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RzH3EcjGsmr2Lb0T:-L81hPkbC4qquq1DHpC0:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L74)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

maybe return ([]byte, bool) is better for client to check if the result is valid.
</blockquote></details>

Done.

---

*[trie/trie.go, line 78 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80RpPc3JxBvaPPL7_U:-L81a19H-pz3QQ4vr-ZQ:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L78)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> val != nil &&

len(val) is enough, len(nil) is 0
</blockquote></details>

Done.

---

*[trie/trie.go, line 99 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80SaxZ6mqVByGbWp_A:-L81_VR3-gF8yTSFPZJ1:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L99)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> common.Hash{}

common.EmptyHash
</blockquote></details>

Done.

---

*[trie/trie.go, line 103 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80SjivBtKi2l9jWV3a:-L81Z4xwC7gb-Xofc7oO:b5vcv05) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L103)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

The code below is the same with previous API, copy mistake? If not, please avoid copy-paste.
</blockquote></details>

No. Hash function just return hash value and not change the database. It may be used by TxHash of blockhead. run just one time.
Commit function will save to db. It may be used by Statedb of blockhead.  

---

*[trie/trie.go, line 113 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80vIrU1CZ2eh75DB16:-L81YQi58lRK3cUn4iQ7:blupw66) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L113)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

unnecessary buf & sha parameter, please refer to crypto.Keccak256Hash for hash calculation.
</blockquote></details>

Keccak256Hash will new a keccak256 hasher every time call hash.not suitable for this case. 

---

*[trie/trie.go, line 121 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80vmQMBQad4egdDgQb:-L81hXDZ0b28pmwlBSdh:bgjmjv1) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L121)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Noder interface has Hash() API already, why not implement it in all kinds of Node? 
</blockquote></details>

It just returns the hash value for caller 

---

*[trie/trie.go, line 179 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80UDZU79lVDhxBWW3p:-L81oO4b1R-0MwSkaJ13:b51e6kh) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L179)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Suggest to add necessary comments for such long method.
</blockquote></details>

I have split the function and  add some comments

---

*[trie/trie.go, line 181 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80UQBS4YXv9p-Bijpo:-L81i07T1AcsRepQnwDE:b5bgzh2) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L181)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

How about use polymorphsim to simplify the code, since such long method is really bad to read and hard to understand in a short time.
</blockquote></details>

do it later:)

---

*[trie/trie.go, line 378 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80VdSnE_sRFEftiOyv:-L81XL7O7qubtEguiDqc:bb7f2mf) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L378)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> val == nil

len(val) == 0
</blockquote></details>

done

---

*[trie/trie.go, line 393 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80VqmE6k0YTLHyCj_O:-L81iEBT47uh8cPZfm2D:bagrr2z) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L393)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

error handling?
</blockquote></details>

error should be handled by  default switch

---

*[trie/trie_test.go, line 26 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80yD170g_4qr1ilfG-:-L81jVfs7eHt2GuvnKnb:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L26)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

delete the temp folder before panic.
</blockquote></details>

Done.

---

*[trie/trie_test.go, line 52 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80zLrRD-LtwnVJ8r1G:-L81jfxC4KOqgK_HZsKQ:b-nabbfv) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L52)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

I think it's better to assert the trie structure (different nodes/values), instead of the Get API. Or add more tests about the data structure.
</blockquote></details>

add more tests later:)

---

*[trie/trie_test.go, line 76 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80yyxxD8cEBjKkvBhK:-L81kG210b-AgVy-mAOf:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L76)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Any assertion after commit? If no, it's unnecessary to commit the updates.
</blockquote></details>

Done.

---

*[trie/trie_test.go, line 95 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81-0PN3W3-Sn4yuiDr:-L81kvqW30IbTcz6d_tG:bn0ccbw) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie_test.go#L95)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Above Update statements are duplicated with previous test, suggest to define common test data.

And again, suggest to assert the data structure.
</blockquote></details>

change some test data; add more tests later:)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-:-L81nbd071QlcvT4-2qw:b8r5d1a)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[trie/node.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81xSBx1GRp-SxERpSe:-L81xSBx1GRp-SxERpSf:b-5szf4m) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L29)):*
> ```Go
> 	Node
> 	Key      []byte // for shared nibbles
> 	Nextnode noder  // for next node
> ```

Is it possible to declare the Nextnode as BranchNode type for type safety?

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L81zwi7EyycZXfOHQ-l:b-ufsozg) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
<details><summary><i>Previously, jessonzheng wrote…</i></summary><blockquote>

prefix is used by different tree;such as account state tree,  contract state tree. they all use 32 bytes hash as key.  
</blockquote></details>

This is an exported API, and MPT client may be not aware the prefix concept. So, how about rename it to key or add necessary comments for the input parameters?

---

*[trie/trie.go, line 107 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L82-u1JC8Ydpb8O-Lx2:-L82-u1JC8Ydpb8O-Lx3:bgx3y3s) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L107)):*
> ```Go
> 	if t.root != nil {
> 		buf := new(bytes.Buffer)
> 		sha := sha3.NewKeccak256()
> ```

Use common hash method under crypto pkg?

If you want to cache the sha obj, how about use one sha instance for the whole tree?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 12 unresolved discussions.

---

*[trie/trie.go, line 6 at r3](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L82b9whVHfIWFhbDHxt:-L82b9whVHfIWFhbDHxu:b-xdb1v) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L6)):*
> package trie

just curious, can this trie resist second preimage attack?
or just leave it to caller

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 13 unresolved discussions.

---

*[trie/node.go, line 10 at r3](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L82j5RhNzFtAAcLjrKO:-L82j5RhNzFtAAcLjrKP:b83m760) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L10)):*
> high

height

another question, for small number of nodes, obviously,17 is not the optimal choice, the path is shorter, however, the number of leaf nodes will be larger, so the total times of comparing will be larger.
how can you prove the number 17 is best value, could you please some links?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

jessonzheng:
Review status: 0 of 3 files reviewed at latest revision, 13 unresolved discussions.

---

*[trie/node.go, line 1 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L7xItKwAzHc7ybqKrbS:-L86-3GP9sjLN61XGOeP:bem8ybu) ([raw file](https://github.com/seeleteam/go-seele/blob/f0966d8ce801d74832d030eaf86a1d258b92aa27/trie/node.go#L1)):*
<details><summary><i>Previously, jessonzheng wrote…</i></summary><blockquote>

Done.
</blockquote></details>

should do some benchmark later

---

*[trie/node.go, line 10 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L82j5RhNzFtAAcLjrKO:-L85zzYe9DcaABPitpwf:b-8ssps4) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L10)):*
> height

---

*[trie/node.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81xSBx1GRp-SxERpSe:-L85bksgADczKvCyWaF3:bl6wjjx) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L29)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Is it possible to declare the Nextnode as BranchNode type for type safety?
</blockquote></details>

No,Nextnode maybe leafnode

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L862V3j6jOixPyWIOlq:b-cueeas) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

This is an exported API, and MPT client may be not aware the prefix concept. So, how about rename it to key or add necessary comments for the input parameters?
</blockquote></details>

add comments for param prefix 

---

*[trie/trie.go, line 6 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L82b9whVHfIWFhbDHxt:-L860Mxj9l8Z4nteqYeP:b38d43g) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L6)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> package trie

just curious, can this trie resist second preimage attack?
or just leave it to caller
</blockquote></details>

leave it to caller;hash key normally is common.address

---

*[trie/trie.go, line 107 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L82-u1JC8Ydpb8O-Lx2:-L8642hMAzNe4AMCJ3o3:b-lba24e) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L107)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Use common hash method under crypto pkg?

If you want to cache the sha obj, how about use one sha instance for the whole tree?
</blockquote></details>

move sha in trie struct 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 13 unresolved discussions.

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L85AjLN8LCQdTtPhQ_B:b-e1dg2x) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

This is an exported API, and MPT client may be not aware the prefix concept. So, how about rename it to key or add necessary comments for the input parameters?
</blockquote></details>

Maybe the right name is dbName?

Moreover, strongly recommend to add necessary comments about the input parameters

---

*[trie/trie.go, line 51 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85BOhS1SnTgyDoBTEB:-L85BOhS1SnTgyDoBTEC:b-4r6d1w) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L51)):*
> Update

According to the implementation, it's AddOrUpdate, not only Update

---

*[trie/trie.go, line 84 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85Cy8UEC4LjGA1OFnd:-L85Cy8UEC4LjGA1OFne:b-cgzaxv) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L84)):*
> ```Go
> 		return nil, false
> 	}
> 	val, newroot := t.get(t.root, key, 0)
> ```

If the root is null, why still try to get the value by key? From the implementation, the t.get(nil, key, 0) always return (nil, nil), so below codes are useless.

---

*[trie/trie.go, line 118 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85Dgt82gsnIcgkaAVD:-L85Dgt82gsnIcgkaAVE:b-3ryen5) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L118)):*
> node.IsDirty() == false

!node.IsDirty()

---

*[trie/trie.go, line 180 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85FGmc6xWg0x9Yo_XV:-L85FGmc6xWg0x9Yo_XW:ba2s5m1) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L180)):*
> ```Go
> }
> 
> func (t *Trie) insert(node noder, key []byte, value []byte) (bool, noder, error) {
> ```

suggest to add necessary comments about he return value. I was thinking that the 1st return value is used to indicate whether insertion succeed. But from your code, it's "dirty"

---

*[trie/trie.go, line 187 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85zhnu-HFiJnHBwLsZ:-L85EEJ70SbjxVPTfT49:by56q5g) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L187)):*
> ```Go
> 		return t.insertLeafNode(n, key, value)
> 	case *BranchNode:
> 		_, child, _ := t.insert(n.Children[key[0]], key[1:], value)
> ```

Error never occur?

---

*[trie/trie.go, line 523 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85wxqO-cKCViwP6II7:-L85wxqO-cKCViwP6II8:bp5qcgd) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L523)):*
> byte(numBranchNodes-1)

I think 16 is better to indicate the hex concept. However, recommend to use:
  b >> 4
  b & 0x0F

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions.

---

*[trie/node.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81xSBx1GRp-SxERpSe:-L866xQB4ILJ2750NTgh:bioj7x4) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L29)):*
<details><summary><i>Previously, jessonzheng wrote…</i></summary><blockquote>

No,Nextnode maybe leafnode
</blockquote></details>

If the nextnode is leafnode, then this ExtendNode could be merged with the leaf node to a new leaf node.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions.

---

*[trie/node.go, line 57 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L867bESBlZroMWpAJtX:-L867bESBlZroMWpAJtY:b-w9xcwc) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L57)):*
> Hash()

coding style: 
one line for the following functions,
func (n hashNode) Hash() []byte { return n }
func (n hashNode) IsDirty() bool { return false }

---

*[trie/trie.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L868Fzy8UHHuS4BGSNb:-L868Fzy8UHHuS4BGSNc:bpadi68) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L29)):*
> //prefix

one blank between '//' and 'prefix'

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[trie/trie.go, line 262 at r4](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L86A9Au2CRBcdqp2niX:-L86A9Au2CRBcdqp2niY:b-xsob7p) ([raw file](https://github.com/seeleteam/go-seele/blob/2d62b1cc1d270932b539e65faf575d22ca20268f/trie/trie.go#L262)):*
> ```Go
> func (t *Trie) insertLeafNode(n *LeafNode, key []byte, value []byte) (bool, noder, error) {
> 	matchlen := matchkeyLen(n.Key, key)
> 	if matchlen == len(n.Key) { // key match, change the value of leaf node
> ```

As we discussed, cannot update the existing record, but add a new one and return it back.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[trie/trie.go, line 33 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86Ct9fCKeRqRn7OSxt:-L86Ct9fCKeRqRn7OSxu:b-jh8abi) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L33)):*
> NewTrie

I suppose when create a new trie, there must be at least one node.
According to current implementation, just pass a root hash, which is calculated by caller from the nodes in db, right?
if so, how can we make sure the hash value is matched with that of values?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Review status: 0 of 3 files reviewed at latest revision, 17 unresolved discussions.

---

*[trie/trie.go, line 523 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85wxqO-cKCViwP6II7:-L86EDuMBCAOXY46vNkY:b5szjlh) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L523)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> byte(numBranchNodes-1)

I think 16 is better to indicate the hex concept. However, recommend to use:
  b >> 4
  b & 0x0F
</blockquote></details>

and nice to add some convert samples as comments for this function to make code more understandable.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[trie/trie.go, line 344 at r4](https://reviewable.io:443/reviews/seeleteam/go-seele/85#-L86HDANSaW-J9hOoEIV:-L86HDANSaW-J9hOoEIW:b-1v1ky0) ([raw file](https://github.com/seeleteam/go-seele/blob/2d62b1cc1d270932b539e65faf575d22ca20268f/trie/trie.go#L344)):*
> ```Go
> 				}
> 			}
> 			switch childnode := childnode.(type) {
> ```

case *BranchNode missed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/seeleteam/go-seele/85)*

jessonzheng:
Review status: 0 of 3 files reviewed at latest revision, 18 unresolved discussions.

---

*[trie/node.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L81xSBx1GRp-SxERpSe:-L86CIIvEdrBzaDRdfqf:buc33aw) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L29)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

If the nextnode is leafnode, then this ExtendNode could be merged with the leaf node to a new leaf node.
</blockquote></details>

??
 nextnode maybe leafnode or branchenode or hashnode.... It should be a noder interface

---

*[trie/node.go, line 57 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L867bESBlZroMWpAJtX:-L86CyOTCTERbi8vemUw:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/node.go#L57)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> Hash()

coding style: 
one line for the following functions,
func (n hashNode) Hash() []byte { return n }
func (n hashNode) IsDirty() bool { return false }


</blockquote></details>

Done.

---

*[trie/trie.go, line 33 at r2](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L80OKEcAc0VVnne9wPO:-L86GjZw1hTSVzoH1R1R:b-b78x1x) ([raw file](https://github.com/seeleteam/go-seele/blob/502b6c2ec982eda473ceaef2c76ef0afe48e049c/trie/trie.go#L33)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Maybe the right name is dbName?

Moreover, strongly recommend to add necessary comments about the input parameters
</blockquote></details>

change it to dbprefix

---

*[trie/trie.go, line 29 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L868Fzy8UHHuS4BGSNb:-L868p0S7NmpBBz3nj5m:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L29)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> //prefix

one blank between '//' and 'prefix'
</blockquote></details>

Done.

---

*[trie/trie.go, line 33 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86Ct9fCKeRqRn7OSxt:-L86F5rv3j-S100vP8f1:b-lfze85) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L33)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> NewTrie

I suppose when create a new trie, there must be at least one node.
According to current implementation, just pass a root hash, which is calculated by caller from the nodes in db, right?
if so, how can we make sure the hash value is matched with that of values?

</blockquote></details>

add errNodeNotExist error ,and return nil if hash value is not match

---

*[trie/trie.go, line 51 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85BOhS1SnTgyDoBTEB:-L8687F7EYo6jgHuFVH5:b-bbvm9c) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L51)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> Update

According to the implementation, it's AddOrUpdate, not only Update
</blockquote></details>

replace Update to Put

---

*[trie/trie.go, line 84 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85Cy8UEC4LjGA1OFnd:-L86A1VyB5dj7fz98hg7:b-pi9dd9) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L84)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

If the root is null, why still try to get the value by key? From the implementation, the t.get(nil, key, 0) always return (nil, nil), so below codes are useless.
</blockquote></details>

remove them

---

*[trie/trie.go, line 118 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85Dgt82gsnIcgkaAVD:-L868daa8tqd5jB6RGcR:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L118)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

> node.IsDirty() == false

!node.IsDirty()
</blockquote></details>

Done.

---

*[trie/trie.go, line 180 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85FGmc6xWg0x9Yo_XV:-L86HHIv6J0w_iU1zJcF:b-jzaqoy) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L180)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

suggest to add necessary comments about he return value. I was thinking that the 1st return value is used to indicate whether insertion succeed. But from your code, it's "dirty"
</blockquote></details>

yes.It mean success. It also assign to node.dirty.
add some comments for it

---

*[trie/trie.go, line 187 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85zhnu-HFiJnHBwLsZ:-L86G0i-ENKugnwBZnBk:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L187)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

Error never occur?
</blockquote></details>

Done.

---

*[trie/trie.go, line 523 at r3](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L85wxqO-cKCViwP6II7:-L86Bk76Al1zxTOdORzx:bo76y0s) ([raw file](https://github.com/seeleteam/go-seele/blob/7bf36801edaf02770761ddd41241d4b7fd4d23e1/trie/trie.go#L523)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

and nice to add some convert samples as comments for this function to make code more understandable.
</blockquote></details>

later to change them, or numBranchNodes define lose its value. 
add some comments

---

*[trie/trie.go, line 262 at r4](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86A9Au2CRBcdqp2niX:-L86AqKe5zQu5EcGZmRj:bxdiu1w) ([raw file](https://github.com/seeleteam/go-seele/blob/2d62b1cc1d270932b539e65faf575d22ca20268f/trie/trie.go#L262)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

As we discussed, cannot update the existing record, but add a new one and return it back.
</blockquote></details>

this also will generate new one in db because the hash key changed

---

*[trie/trie.go, line 344 at r4](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86HDANSaW-J9hOoEIV:-L86JmaE40p66mHIdsrE:bqv9fi7) ([raw file](https://github.com/seeleteam/go-seele/blob/2d62b1cc1d270932b539e65faf575d22ca20268f/trie/trie.go#L344)):*
<details><summary><i>Previously, boqiu wrote…</i></summary><blockquote>

case *BranchNode missed
</blockquote></details>

don't do this case; node can't be merged

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

boqiu:
Reviewed 3 of 3 files at r5.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[trie/trie.go, line 81 at r5](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86Rp2n2YJiuvKY8S0f:-L86Rp2n2YJiuvKY8S0g:b-ings6s) ([raw file](https://github.com/seeleteam/go-seele/blob/e9b0d0fb6e74e98a702096f3e457b1d1eac69352/trie/trie.go#L81)):*
> key = keybytesToHex(key)

move this statement into if {} as in Delete().

another suggestion, better to move this logic and t.root != nil into their corresponding internal functions to make the public function more clear.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[trie/trie.go, line 87 at r5](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86SVujDkUtA59d326-:-L86SVujDkUtA59d3260:b1dmlzw) ([raw file](https://github.com/seeleteam/go-seele/blob/e9b0d0fb6e74e98a702096f3e457b1d1eac69352/trie/trie.go#L87)):*
> return nil, false

it's redundant, can be removed.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

jessonzheng:
Review status: 2 of 3 files reviewed at latest revision, 6 unresolved discussions.

---

*[trie/trie.go, line 81 at r5](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86Rp2n2YJiuvKY8S0f:-L87-mW48KnEtix53wPr:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/e9b0d0fb6e74e98a702096f3e457b1d1eac69352/trie/trie.go#L81)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> key = keybytesToHex(key)

move this statement into if {} as in Delete().

another suggestion, better to move this logic and t.root != nil into their corresponding internal functions to make the public function more clear.
</blockquote></details>

Done.

---

*[trie/trie.go, line 87 at r5](https://beta.reviewable.io/reviews/seeleteam/go-seele/85#-L86SVujDkUtA59d326-:-L870-2q2XvLsCx_KcQp:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/e9b0d0fb6e74e98a702096f3e457b1d1eac69352/trie/trie.go#L87)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> return nil, false

it's redundant, can be removed.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

WTer:
Reviewed 2 of 3 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/85)*

PR #114, opened 2018-04-10
fix miner write block failed; close msg channel when peer disconnect

lucusfly:
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/114)

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -308,7 +308,7 @@ func updateStatedb(statedb *state.Statedb, minerRewardTx *types.Transaction, txs
 
 		fromStateObj := statedb.GetOrNewStateObject(tx.Data.From)
 		fromStateObj.SubAmount(tx.Data.Amount)
-		fromStateObj.SetNonce(fromStateObj.GetNonce() + 1)
+		fromStateObj.SetNonce(tx.Data.AccountNonce + 1)
 
 		toStateObj := statedb.GetOrNewStateObject(*tx.Data.To)
 		toStateObj.AddAmount(tx.Data.Amount)
--- a/miner/task.go
+++ b/miner/task.go
@@ -37,7 +37,8 @@ func (task *Task) applyTransactions(seele *seele.SeeleService, statedb *state.St
 	rewardValue := big.NewInt(minerRewardAmount)
 	reward := types.NewTransaction(common.Address{}, seele.Coinbase, rewardValue, 0)
 	reward.Signature = &crypto.Signature{}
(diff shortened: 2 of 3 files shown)


WTer:
Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---

*[core/blockchain.go, line 311 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/114#-L9iK0Ez1BmWgqVtp8q6:-L9iK0Ez1BmWgqVtp8q7:b-86vkab) ([raw file](https://github.com/seeleteam/go-seele/blob/1a42da9c7590d1f21bf6cf41dc1db2abc60be15c/core/blockchain.go#L311)):*
> ```
> tx.Data.AccountNonce
> ```
Suppose accountA transfer 2 txs in different threads, for each tx the nonce may be the same number, the  fromStateObj.GetNonce() is persistent number, but you're using the temp and not accurate number tx.Data.AccountNonce, which will lead to a wrong nonce.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/114)*

boqiu:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[core/blockchain.go, line 311 at r1](https://beta.reviewable.io/reviews/seeleteam/go-seele/114#-L9iK0Ez1BmWgqVtp8q6:-L9iNdcu9u0lFkwyOgMG:bcv8qtd) ([raw file](https://github.com/seeleteam/go-seele/blob/1a42da9c7590d1f21bf6cf41dc1db2abc60be15c/core/blockchain.go#L311)):*
<details><summary><i>Previously, WTer wrote…</i></summary><blockquote>

> ```
> tx.Data.AccountNonce
> ```
Suppose accountA transfer 2 txs in different threads, for each tx the nonce may be the same number, the  fromStateObj.GetNonce() is persistent number, but you're using the temp and not accurate number tx.Data.AccountNonce, which will lead to a wrong nonce.
</blockquote></details>

Any special purpose of using tx nonce to update the state db? user could set uint64.MaxValue as tx nonce (e.g. via API), if so the new account nonce in state db will become to 0. Here I suggest to keep the original logic to restrict the nonce value change.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/114)*

WTer:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/seeleteam/go-seele/114)*

PR #135, opened 2018-05-02
copy evm to seele

boqiu:
EVM integration iteration 1: copy ETH vm code to Seele

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/135)


boqiu:
This is to track the code change based on EVM source code. So, temporarily check in this PR with build error.

PR #144, opened 2018-05-03
fix bugs in node

sufay:
1: in node.Start, services successful started previously will be stopped when a service fails to stop. but in the implementation, no iteration is used.

2: in node.Start, after the p2p server is started,  it should be stopped when any error occurs.  a Stop function should be added in p2p.server

3: in node.Stop, when any service fails to stop, the for loop should store the error and go on rather than return an error.  a StopError struct is needed for the issue. in addition, the running p2p server needs to be terminated too.

@herove @reedhong @lucusfly @chenhuaying @WTer

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/144)


lucusfly:
Hi sufay, we use reviewable.io tool to review code. please click this button above to start review.
image

lucusfly:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/144)*

PR #150, opened 2018-05-05
rename several functions and variables

sufay:
rename several names:
1. store.go and db_store.go: rename HashBlock to HasBlock, which means wether a block exists
2. block_index.go: change findBestBlockHash to findBestBlockIndex, which is intended to find the best block index; 
3. block_index.go: change NewBlockLeaf to NewBlockLeaves so as to comply with the struct name
4. transaction.go: renaming ErrSigMissed to ErrSigMissing  seems to be better
5. blockchain.go: ErrBlockAlreadyExists should be better compared to ErrBlockAlreadyExist

add and modify some comments mainly in core package

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/150)


lucusfly:
Reviewed 19 of 19 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/150)*

PR #151, opened 2018-05-07
add GetDB for trie

jessonzheng:
add getdb function,return database interface of trie. It will be used by statedb later.

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/151)

### DIFF
--- a/trie/trie.go
+++ b/trie/trie.go
@@ -52,6 +52,11 @@ func NewTrie(root common.Hash, dbprefix []byte, db database.Database) (*Trie, er
 	return trie, nil
 }
 
+// GetDB returns database interface of trie 
+func (t *Trie) GetDB() database.Database{
+	return t.db
+}
+
 // Put add or update [key,value] in the trie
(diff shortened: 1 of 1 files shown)


lucusfly:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[trie/trie.go, line 56 at r1](https://reviewable.io/reviews/seeleteam/go-seele/151#-LBu66l_A6YT8_tBjPNi:-LBv5Zs-1IqXO_lKmpU4:b-ukwegr) ([raw file](https://github.com/seeleteam/go-seele/blob/29e8a622678e5d3700cedb76ec0cdb3f3e4f08da/trie/trie.go#L56)):*

Do you want to store any other info in the same DB with trie? If so, how about create the statedb with a DB instance?
</blockquote></details>

could you merge it with the change that use it? this change is no meaning and not enough for a PR.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/151)*

PR #183, opened 2018-05-17
fix bugs in task and event

sufay:
1. miner/task.go:  as a result of miner reward is considered as a transaction,  in the condition clause of task.chooseTransactions, BlockTransactionNumberLimit should be replaced with BlockTransactionNumberLimit - 1.  otherwise,  blockchain.validateBlock will fail if the number of txs is up to the maximum
2. event/manager.go:  in event.Fire, we should remove listeners before they are invoked in case that a listener callback will add itself again. if not, the attempt to register itself once again could fail. because it can occur before the ultimate removing for either uncertainty of goroutine or multiple sync callback
3. event/manager.go:  add lock in removeOnceListener, even though it's only called by Fire.

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)

### DIFF
--- a/event/manager.go
+++ b/event/manager.go
@@ -17,21 +17,23 @@ type EventManager struct {
 	listeners []eventListener
 }
 
-// Fire fires the event and returns it after all listeners have done
-// their jobs.
+// Fire triggers all listeners with the specified event and removes listeners which run only once.
 func (h *EventManager) Fire(e Event) {
 	h.lock.Lock()
 	defer h.lock.Unlock()
-
-	for _, l := range h.listeners {
+    
+	listeners := make([]eventListener, len(h.listeners))
+	copy(listeners, h.listeners)
+	
+	h.removeOnceListener()
+	
+	for _, l := range listeners {
 		if l.IsAsyncListener {
 			go l.Callable(e)
 		} else {
 			l.Callable(e)
 		}
 	}
-
-	h.removeOnceListener()
 }
 
 // AddListener registers a listener.
@@ -101,14 +103,17 @@ func (h *EventManager) RemoveListener(callback EventHandleMethod) {
 
 // removeOnceListener removes all listeners which run only once
 func (h *EventManager) removeOnceListener() {
-	listener := make([]eventListener, 0, len(h.listeners))
+	h.lock.Lock()
+	defer h.lock.Unlock()
+	
+	listeners := make([]eventListener, 0, len(h.listeners))
 	for _, l := range h.listeners {
 		if !l.IsOnceListener {
-			listener = append(listener, l)
+			listeners = append(listeners, l)
 		}
 	}
 
-	h.listeners = listener
+	h.listeners = listeners
 }
 
 // find finds listener existing in the manager
--- a/event/manager.go
+++ b/event/manager.go
@@ -102,10 +102,7 @@ func (h *EventManager) RemoveListener(callback EventHandleMethod) {
 }
 
 // removeOnceListener removes all listeners which run only once
-func (h *EventManager) removeOnceListener() {
-	h.lock.Lock()
-	defer h.lock.Unlock()
-	
+func (h *EventManager) removeOnceListener() {	
 	listeners := make([]eventListener, 0, len(h.listeners))
 	for _, l := range h.listeners {
 		if !l.IsOnceListener {
--- a/event/manager.go
+++ b/event/manager.go
@@ -21,19 +21,16 @@ type EventManager struct {
 func (h *EventManager) Fire(e Event) {
 	h.lock.Lock()
 	defer h.lock.Unlock()
-    
-	listeners := make([]eventListener, len(h.listeners))
-	copy(listeners, h.listeners)
 	
-	h.removeOnceListener()
-	
-	for _, l := range listeners {
+	for _, l := range h.listeners {
 		if l.IsAsyncListener {
 			go l.Callable(e)
 		} else {
 			l.Callable(e)
 		}
 	}
+
+	h.removeOnceListener()
 }
 
 // AddListener registers a listener.
--- a/miner/task.go
+++ b/miner/task.go
@@ -53,7 +53,7 @@ func (task *Task) applyTransactions(seele SeeleBackend, statedb *state.Statedb, 
 }
 
 func (task *Task) chooseTransactions(seele SeeleBackend, statedb *state.Statedb, txs map[common.Address][]*types.Transaction, log *log.SeeleLog) {
-	for i := 0; i < core.BlockTransactionNumberLimit; i++ {
+	for i := 0; i < core.BlockTransactionNumberLimit - 1; i++ {
 		tx := popBestFeeTx(txs)
 		if tx == nil {
 			break


boqiu:
Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[event/manager.go, line 26 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-00aCaN8FGVyNbpA:-LCl-00aCaN8FGVyNbpB:b-3pijl1) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L26)):*
> ```Go
>     
> 	listeners := make([]eventListener, len(h.listeners))
> 	copy(listeners, h.listeners)
> ```

Always copy the listeners will impact the perf.

---

*[event/manager.go, line 107 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCkzGsd279ofvb8sG0W:-LCkzGsd279ofvb8sG0X:badjbvt) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L107)):*
> ```Go
> func (h *EventManager) removeOnceListener() {
> 	h.lock.Lock()
> 	defer h.lock.Unlock()
> ```

this method will be invoked when Fire an event, which also required to Lock(). As a result, deadlock occur.

---

*[miner/task.go, line 56 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-zNsE9FQVdbQmcz6:-LCl-zNsE9FQVdbQmcz7:b16vgs4) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/miner/task.go#L56)):*
> ```Go
> 
> func (task *Task) chooseTransactions(seele SeeleBackend, statedb *state.Statedb, txs map[common.Address][]*types.Transaction, log *log.SeeleLog) {
> 	for i := 0; i < core.BlockTransactionNumberLimit - 1; i++ {
> ```

I think this API only choose tx from customer txs, not including the reward one.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

WTer:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[event/manager.go, line 24 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl12kgAE7d8mq4JC4o:-LCl12kgAE7d8mq4JC4p:bpmm9in) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L24)):*
> ```
> ····
> ```
remove these unnecessary blanks.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

boqiu:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[event/manager.go, line 26 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-00aCaN8FGVyNbpA:-LCl15eh0BpJyA1LR21t:b-f30qfz) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L26)):*

Always copy the listeners will impact the perf.
</blockquote></details>

If our system do not require once listener register itself, then current code works fine. But, copy listeners every time really ugly.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

sufay:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[event/manager.go, line 26 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-00aCaN8FGVyNbpA:-LCn-8Mh295FZFxAO6ln:b-gcnz4b) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L26)):*

If our system do not require once listener register itself, then current code works fine. But, copy listeners every time really ugly.
</blockquote></details>

a copy  is needed for that purpose(allow to registered again in callback), or else  a newly registered listener(appended to the end) will be fired again for interation. 

---

*[event/manager.go, line 107 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCkzGsd279ofvb8sG0W:-LCn0avH8miXtbHRnlxU:b-hezygn) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L107)):*

this method will be invoked when Fire an event, which also required to Lock(). As a result, deadlock occur.
</blockquote></details>

I'm sorry. i will remove it

---

*[miner/task.go, line 56 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-zNsE9FQVdbQmcz6:-LCn1u4EDVOsl9yHR1Bq:b-lyyceu) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/miner/task.go#L56)):*

I think this API only choose tx from customer txs, not including the reward one.
</blockquote></details>

just because chooseTransactions chooses txs from tx pool, so here max number is LIMIT - 1. otherwise , suppose LIMIT  txs are chosen from tx pool(possibly),  the total number including reward will be LIMIT +1 so that IF clause "len(block.Transactions) > LIMIT" will be true  in validateBlock.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

sufay:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[event/manager.go, line 24 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl12kgAE7d8mq4JC4o:-LCn86iU00R8IjffpyWQ:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L24)):*
<details><summary><i>Previously, WTer (Wensi Liu) wrote…</i></summary><blockquote>

> ```
> ····
> ```
remove these unnecessary blanks.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

lucusfly:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

boqiu:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[event/manager.go, line 26 at r1](https://reviewable.io/reviews/seeleteam/go-seele/183#-LCl-00aCaN8FGVyNbpA:-LCxXKssExZ7FeYNEZYs:b-o3agq6) ([raw file](https://github.com/seeleteam/go-seele/blob/3edd8c8ca4fb0b1a2c332b9e832e92a8a284558c/event/manager.go#L26)):*

a copy  is needed for that purpose(allow to registered again in callback), or else  a newly registered listener(appended to the end) will be fired again for interation. 
</blockquote></details>

The once listener is not used in Seele (now, it's over designed), it is unnecessary to copy the listeners every time when event fired, which will impact the perf of normal event handling. On the other hand, non-once listener do not require to register itself during event handling.

As for the implementation of once listener, worse bug occur before the listeners copy. That's deadlock, because both Fire and addEventListener APIs require to hold the lock. To support once listener handling reg itself again, I think we could introduce listener state or use 2 listener arrays for normal and nonce listeners respectively.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

boqiu:
Reviewed 1 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

WTer:
Reviewed 1 of 2 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/183)*

PR #188, opened 2018-05-20
fix issues in task

sufay:
two issues in task.chooseTransactions:
1.  when a tx can't be validated or applied, the loop counter should make no iteration to collect as many txs as possible.
2.  for blockchain.ApplyTransaction, the parameter txIndex refers to the index of a tx in a block. so to keep this meaning, add 1 to the current value because the reward tx is located in the index 0. Although this will not affect execution.

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/188)

### DIFF
--- a/miner/task.go
+++ b/miner/task.go
@@ -53,7 +53,7 @@ func (task *Task) applyTransactions(seele SeeleBackend, statedb *state.Statedb, 
 }
 
 func (task *Task) chooseTransactions(seele SeeleBackend, statedb *state.Statedb, txs map[common.Address][]*types.Transaction, log *log.SeeleLog) {
-	for i := 0; i < core.BlockTransactionNumberLimit - 1; i++ {
+	for i := 0; i < core.BlockTransactionNumberLimit - 1; {
 		tx := popBestFeeTx(txs)
 		if tx == nil {
 			break
@@ -67,7 +67,7 @@ func (task *Task) chooseTransactions(seele SeeleBackend, statedb *state.Statedb,
 			continue
 		}
 
-		receipt, err := seele.BlockChain().ApplyTransaction(tx, i, seele.GetCoinbase(), statedb, task.header)
+		receipt, err := seele.BlockChain().ApplyTransaction(tx, i+1, seele.GetCoinbase(), statedb, task.header)
 		if err != nil {
 			log.Error("apply tx failed, %s", err.Error())
 			continue
@@ -75,6 +75,8 @@ func (task *Task) chooseTransactions(seele SeeleBackend, statedb *state.Statedb,
 
 		task.txs = append(task.txs, tx)
 		task.receipts = append(task.receipts, receipt)
+		
+		i++
 	}
 }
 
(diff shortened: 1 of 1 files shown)


boqiu:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[miner/task.go, line 70 at r1](https://reviewable.io/reviews/seeleteam/go-seele/188#-LD-S9eY3k04UCP05im7:-LD-S9eY3k04UCP05im8:bai28zt) ([raw file](https://github.com/seeleteam/go-seele/blob/0b6d2d43952ea993adb632455f19cdcc5a4d4874/miner/task.go#L70)):*
> ```Go
> 		}
> 
> 		receipt, err := seele.BlockChain().ApplyTransaction(tx, i+1, seele.GetCoinbase(), statedb, task.header)
> ```

In blockchain, the index starts from 0, and not including the reward tx. Please refer to the updateStateDB method in blockchain.go

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/188)*

boqiu:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[miner/task.go, line 70 at r1](https://reviewable.io/reviews/seeleteam/go-seele/188#-LD-S9eY3k04UCP05im7:-LD1vb_I2DnGyNsEcCqN:byrteal) ([raw file](https://github.com/seeleteam/go-seele/blob/0b6d2d43952ea993adb632455f19cdcc5a4d4874/miner/task.go#L70)):*

in db_store.putBlockInternal, we can see the final storage rule in blockchain: the tx index of reward is 0. there is also a TxIndex struct which indicates the same. Although it's a bit different from that in statedb(in statedb, txIndex is intended for handling receipt). But I think we should keep consistency globally to avoid confusion?
</blockquote></details>

Yes, you're right, thanks to point out such issue. Let's keep the behavior in this file, since reward tx do not require a receipt. As for the issue in blockchain store, we will fix it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/188)*

sufay:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[miner/task.go, line 70 at r1](https://reviewable.io/reviews/seeleteam/go-seele/188#-LD-S9eY3k04UCP05im7:-LD20PT9-kH2Gzr_6DRq:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/0b6d2d43952ea993adb632455f19cdcc5a4d4874/miner/task.go#L70)):*

Yes, you're right, thanks to point out such issue. Let's keep the behavior in this file, since reward tx do not require a receipt. As for the issue in blockchain store, we will fix it.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/188)*

boqiu:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/188)*

PR #206, opened 2018-05-24
Add shard in seele module

qiaozhigang:
Create different handle message function for other shard peers, that will handle no message
Broadcast transactions accroding to shardid. 
Shardid of transaction is determined by it from address

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/206)

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -14,6 +14,9 @@ import (
 	"reflect"
 	"sync"
 
+	"github.com/seeleteam/go-seele/p2p/discovery"
+
+	"github.com/seeleteam/go-seele/common"
 	"github.com/seeleteam/go-seele/log"
 	"github.com/seeleteam/go-seele/p2p"
 	"github.com/seeleteam/go-seele/rpc"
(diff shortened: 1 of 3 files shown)


lucusfly:
Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---

*[seele/seeleprotocol.go, line 306 at r1](https://reviewable.io/reviews/seeleteam/go-seele/206#-LDF8NCDBNz_O2sADOmN:-LDF8NCDBNz_O2sADOmO:brgd4s9) ([raw file](https://github.com/seeleteam/go-seele/blob/9a1b92b6bb00228da02d41d3acc4afa0f8909f10/seele/seeleprotocol.go#L306)):*
> ```Go
> }
> 
> func (p *SeeleProtocol) handleMsg(peer *peer) {
> ```

This is method will not be called for other shard peer as you returned when addPeer

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/206)*

PR #232, opened 2018-06-02
fix timestamp issue in transaction

sufay:
1. timestamp is meaningless in a common transaction, as it can be set at will. Set to 0 to filter a transaction by hash more quickly in txpool.AddTransaction(different timestamps will result in different hashes in case other fields are same)
2. as a result of the reason above, two reward txs may be of the same hash if timestamp is same. we can use the block timestamp which increases by height. so, the hashes of reward txs are different in a single chain.

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -371,8 +371,8 @@ func (bc *Blockchain) validateMinerRewardTx(block *types.Block) (*types.Transact
 	if err := bc.engine.ValidateRewardAmount(block.Header.Height, minerRewardTx.Data.Amount); err != nil {
 		return nil, err
 	}
-	 
-	if timestamp := minerRewardTx.Data.Timestamp; timestamp == nil || timestamp.Cmp(block.Header.CreateTimestamp) != 0 {
+	
+	if minerRewardTx.Data.Timestamp != block.Header.CreateTimestamp.Uint64() {
 	    return nil, types.ErrTimestampMismatch
 	}
 
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -371,6 +371,10 @@ func (bc *Blockchain) validateMinerRewardTx(block *types.Block) (*types.Transact
 	if err := bc.engine.ValidateRewardAmount(block.Header.Height, minerRewardTx.Data.Amount); err != nil {
 		return nil, err
 	}
+	 
+	if timestamp := minerRewardTx.Data.Timestamp; timestamp == nil || timestamp.Cmp(block.Header.CreateTimestamp) != 0 {
+	    return nil, types.ErrTimestampMismatch
+	}
 
 	return minerRewardTx, nil
 }
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -91,7 +91,7 @@ func newTestBlock(bc *Blockchain, parentHash common.Hash, blockHeight, txNum, st
 	common.IsShardDisabled = true
 
 	minerAccount := newTestAccount(pow.GetReward(blockHeight), 0)
-	rewardTx, _ := types.NewTransaction(common.Address{}, minerAccount.addr, minerAccount.data.Amount, big.NewInt(0), minerAccount.data.Nonce)
+	rewardTx, _ := types.NewRewardTransaction(minerAccount.addr, minerAccount.data.Amount, big.NewInt(1))
 	rewardTx.Sign(minerAccount.privKey)
 
 	txs := []*types.Transaction{rewardTx}
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -91,8 +91,7 @@ func newTestBlock(bc *Blockchain, parentHash common.Hash, blockHeight, txNum, st
 	common.IsShardDisabled = true
 
 	minerAccount := newTestAccount(pow.GetReward(blockHeight), 0)
-	rewardTx, _ := types.NewRewardTransaction(minerAccount.addr, minerAccount.data.Amount, big.NewInt(1))
-	rewardTx.Sign(minerAccount.privKey)
+	rewardTx, _ := types.NewRewardTransaction(minerAccount.addr, minerAccount.data.Amount, uint64(1))
 
 	txs := []*types.Transaction{rewardTx}
 	for i := uint64(0); i < txNum; i++ {
--- a/core/store/store_test.go
+++ b/core/store/store_test.go
@@ -87,16 +87,17 @@ func Test_blockchainDatabase_Header(t *testing.T) {
 func newTestTx() *types.Transaction {
 	tx := &types.Transaction{
 		Data: &types.TransactionData{
-			From:    *crypto.MustGenerateRandomAddress(),
-			To:      crypto.MustGenerateRandomAddress(),
-			Amount:  big.NewInt(3),
-			Fee:     big.NewInt(0),
-			Payload: make([]byte, 0),
+			From:      *crypto.MustGenerateRandomAddress(),
+			To:        crypto.MustGenerateRandomAddress(),
+			Amount:    big.NewInt(3),
+			Fee:       big.NewInt(0),
+			Timestamp: big.NewInt(0),
+			Payload:   make([]byte, 0),
 		},
 		Signature: &crypto.Signature{[]byte("test sig")},
 	}
 
-	tx.Hash = crypto.MustHash(tx)
+	tx.Hash = crypto.MustHash(tx.Data)
 
 	return tx
 }
--- a/core/store/store_test.go
+++ b/core/store/store_test.go
@@ -87,11 +87,11 @@ func Test_blockchainDatabase_Header(t *testing.T) {
 func newTestTx() *types.Transaction {
 	tx := &types.Transaction{
 		Data: &types.TransactionData{
-			From:      *crypto.MustGenerateRandomAddress(),
-			To:        crypto.MustGenerateRandomAddress(),
-			Amount:    big.NewInt(3),
-			Fee:       big.NewInt(0),
-			Payload:   make([]byte, 0),
+			From:    *crypto.MustGenerateRandomAddress(),
+			To:      crypto.MustGenerateRandomAddress(),
+			Amount:  big.NewInt(3),
+			Fee:     big.NewInt(0),
+			Payload: make([]byte, 0),
 		},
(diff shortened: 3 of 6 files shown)


boqiu:
Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---

*[core/blockchain.go, line 375 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3CyxN-xDM32gHVZa8:-LE3CyxN-xDM32gHVZa9:bxb2e0u) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/blockchain.go#L375)):*
> ```Go
> 	}
> 	 
> 	if timestamp := minerRewardTx.Data.Timestamp; timestamp == nil || timestamp.Cmp(block.Header.CreateTimestamp) != 0 {
> ```

timestamp should not be nil. 

On the other hand, we could use uint64 type for timestamp in block header, since it's RLP serializable as well.

---

*[core/store/store_test.go, line 100 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3Da2XDmYDXi83OHYq:-LE3Da2XDmYDXi83OHYr:b-kl8g3z) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/store/store_test.go#L100)):*
> ```Go
> 	}
> 
> 	tx.Hash = crypto.MustHash(tx.Data)
> ```

good catch.

---

*[core/types/transaction.go, line 67 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3Did30xKKycC_4PZb:-LE3Did30xKKycC_4PZc:bhdfmi) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L67)):*
> ```Go
> 	AccountNonce uint64          // AccountNonce is the nonce of the sender account
> 	Fee          *big.Int        // Transaction Fee
> 	Timestamp    *big.Int        // Timestamp is used for the miner reward transaction, referring to the block timestamp
> ```

uint64 is easy to use, please keep the original type.

---

*[core/types/transaction.go, line 93 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3FLt9ECXgLorGOJUs:-LE3FLt9ECXgLorGOJUt:b2y421l) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L93)):*
> ```Go
> // panic if the amount is nil or negative.
> func NewTransaction(from, to common.Address, amount *big.Int, fee *big.Int, nonce uint64) (*Transaction, error) {
> 	tx, err := newTx(from, &to, amount, fee, nonce, nil, big.NewInt(0))
> ```

tx creation timestamp info missed. why not use time.Now()?

In fact, only the reward tx need a special timestamp, so as to distinguish the tx hash and make the timestamp validatable.

---

*[core/types/transaction.go, line 100 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3FTstCK91zD1RjA2k:-LE3FTstCK91zD1RjA2l:b-kuxott) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L100)):*
> ```Go
> }
> 
> func newTx(from common.Address, to *common.Address, amount *big.Int, fee *big.Int, nonce uint64, payload []byte, timestamp *big.Int) (*Transaction, error) {
> ```

Suggest not to change this API, since only reward tx may change the timestamp (to avoid tx hash confliction). If so, just explicitly set the rewardTx.Timestamp = header.Timestamp

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)*

sufay:
Review status: 0 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[core/blockchain.go, line 375 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3CyxN-xDM32gHVZa8:-LEAM3w7AOfCKRSc32hw:b8dp5c6) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/blockchain.go#L375)):*

timestamp should not be nil. 

On the other hand, we could use uint64 type for timestamp in block header, since it's RLP serializable as well.
</blockquote></details>

Done.  yes,  uint64 works well.

---

*[core/types/transaction.go, line 67 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3Did30xKKycC_4PZb:-LEAMSFZDC6AlYMldjfg:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L67)):*

uint64 is easy to use, please keep the original type.
</blockquote></details>

Done.

---

*[core/types/transaction.go, line 93 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3FLt9ECXgLorGOJUs:-LEAMi4c9W8o8c-fYktE:bs2pxjs) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L93)):*

tx creation timestamp info missed. why not use time.Now()?

In fact, only the reward tx need a special timestamp, so as to distinguish the tx hash and make the timestamp validatable.
</blockquote></details>

1. timestamp field makes no sense in a transaction except for reward tx. Because anyone can assign an arbitrary value to it, then broadcast to network. But we can't validate it.  
2. suppose timestamp varies, hashes of transactions will be different when other fields including nonce are same(someone can send a number of transactions like this). If so, in tx_pool.AddTransaction, we can't quickly filter these transactions according to hash.

---

*[core/types/transaction.go, line 100 at r1](https://reviewable.io/reviews/seeleteam/go-seele/232#-LE3FTstCK91zD1RjA2k:-LEAMYyf2JwwO6aeqIjZ:b-896fix) ([raw file](https://github.com/seeleteam/go-seele/blob/aaf68ccb02adccb4e07d35a357310d7d34ca1838/core/types/transaction.go#L100)):*

Suggest not to change this API, since only reward tx may change the timestamp (to avoid tx hash confliction). If so, just explicitly set the rewardTx.Timestamp = header.Timestamp
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)*

boqiu:
Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---

*[core/types/transaction.go, line 120 at r2](https://reviewable.io/reviews/seeleteam/go-seele/232#-LED75ptB0d6Rj_TNFf1:-LED75ptB0d6Rj_TNFf2:b-uec5d9) ([raw file](https://github.com/seeleteam/go-seele/blob/1b796e6b0029914a745d7447fd9083a995d7616d/core/types/transaction.go#L120)):*
> ```Go
> 		Amount:       new(big.Int).Set(amount),
> 		Fee:          new(big.Int).Set(fee),
> 		Timestamp:    uint64(time.Now().UnixNano()),
> ```

I think here we could keep the original logic, that set tx timestamp to time.Now(). Please see my comments as following:

If user want to know the tx creation time, he/she will not change it to any value. Additionally, the timestamp cannot be modified during tx/block broadcast, since the tx signature is calcuated based on timestamp.

As for filter tx by hash in txpool, it has nothing to do with the timestamp value, since the timestamp value will not impact the tx Hash calcualtion perf.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)*

sufay:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[core/types/transaction.go, line 120 at r2](https://reviewable.io/reviews/seeleteam/go-seele/232#-LED75ptB0d6Rj_TNFf1:-LEDaN2i8OkSMSvx0Rzo:b-sb2ent) ([raw file](https://github.com/seeleteam/go-seele/blob/1b796e6b0029914a745d7447fd9083a995d7616d/core/types/transaction.go#L120)):*

I think here we could keep the original logic, that set tx timestamp to time.Now(). Please see my comments as following:

If user want to know the tx creation time, he/she will not change it to any value. Additionally, the timestamp cannot be modified during tx/block broadcast, since the tx signature is calcuated based on timestamp.

As for filter tx by hash in txpool, it has nothing to do with the timestamp value, since the timestamp value will not impact the tx Hash calcualtion perf.
</blockquote></details>

i don't mean that timestamp can be tempered(this is impossible due to signature).  My point is that it can be arbitrarily assigned(say 1 or 2, eg.  which are meaningless to refer to ) before sign and broadcast a transaction.  

For instance, I make a legal transaction with timestamp set to 100 in my local app, then sign and send it to node. Finally it occurs in the canonical blockchain. But the timestamp of the transaction cannot be used to indicate the creation time correctly.

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)*

boqiu:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[core/types/transaction.go, line 120 at r2](https://reviewable.io/reviews/seeleteam/go-seele/232#-LED75ptB0d6Rj_TNFf1:-LEF5tIaB9SguXzSOzfA:bz2ouw7) ([raw file](https://github.com/seeleteam/go-seele/blob/1b796e6b0029914a745d7447fd9083a995d7616d/core/types/transaction.go#L120)):*

i don't mean that timestamp can be tempered(this is impossible due to signature).  My point is that it can be arbitrarily assigned(say 1 or 2, eg.  which are meaningless to refer to ) before sign and broadcast a transaction.  

For instance, I make a legal transaction with timestamp set to 100 in my local app, then sign and send it to node. Finally it occurs in the canonical blockchain. But the timestamp of the transaction cannot be used to indicate the creation time correctly.
</blockquote></details>

Yes, I got your point. Anyway, either set the timestamp to 0 or time.Now() is ok. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/232)*

PR #260, opened 2018-06-08
add stateless check for txs of different shard

sufay:
when a tx with different shard number is sent to a node from client, a state independent check should be performed before sending it to corresponding peer sets for network performance

---
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/260)


sufay:
why has travis-ci not reported build status? what happened to the PR?

PR #427, opened 2018-08-01
add hashlock in util and unit test for hashlock

yanllearnn:
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/427)


yanllearnn:
close for doing it in contract

PR #857, opened 2018-12-04
Fix the low level logs rpc display error

wangff15386:
This change is [Reviewable](https://reviewable.io/reviews/seeleteam/go-seele/857)


wangff15386:
remove more

PR #886, opened 2018-12-23
add another getMiningTarget function

shawnxin89:
It's used after the miner has found the nonce such that sha256(blockheaderhash + nonce) < firstMiningTarget. The mined block will not be broadcasted unless sha256(nonce + coinbase) < secondMiningTarget.

### DIFF
--- a/consensus/pow/engine.go
+++ b/consensus/pow/engine.go
@@ -151,13 +151,13 @@ func getMiningTarget(difficulty *big.Int) *big.Int {
 
 func getSecondMiningTarget(time uint64, parentHeader *types.BlockHeader) *big.Int {
     // target = maxUint256 / current difficulty
-    // current difficulty = 20,000,000 / min((current time - parentTime) / 20, 100)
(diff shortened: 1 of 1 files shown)


consensus/pow/engine.go
```diff
@@ -144,3 +144,28 @@ func verifyTarget(header *types.BlockHeader) error {
 func getMiningTarget(difficulty *big.Int) *big.Int {
 	return new(big.Int).Div(maxUint256, difficulty)
 }
+
+// returns the second mining target based on the difference of current clock time
+// and the timestamp of parent block. Initially, the difficulty should be high
+// This function is used to determine which coinbase can mine.
+
+func getSecondMiningTarget(time uint64, parentHeader *types.BlockHeader) *big.Int {
```
boqiu: Is it designed to resolve the sharding security issue? If yes, please discuss with other team members firstly, we should ensure the solution is doable before any code merge.

PR #904, opened 2019-01-24
change default gas limit to 3000000, nonce to current account nonce

avadesian:
To solve this problem :
when use cmd/vm to create contract, the command can only be executed when the account's balance is set to MaxUint64


avadesian:
The  nonce value should not be smaller than current nonce, therefore accountNonce is OK, of course accountNonce +1 is also OK.
Refer to Line 332 in types/transaction.go
     ` if accountNonce := statedb.GetNonce(tx.Data.From); tx.Data.AccountNonce < accountNonce {`

PR #992, opened 2019-07-11
fix lightchain sync bugs (updatedAncestor, malicious response, blockNumArray)

LincolnZeng:


light/peer.go
```diff
@@ -32,6 +32,7 @@ var (
 	errModeNotMatch    = errors.New("server/client mode mismatch")
 	errGenesisNotMatch = errors.New("genesis hash mismatch")
 	errBlockNotFound   = errors.New("block not found")
+	errBlocksExist     = errors.New("peer sync hashArray already exit locally")
```
shawnxin89: exist?

PR #1031, opened 2019-10-31
Lincoln zeng patch 1

LincolnZeng:


troytrojan01:
Mining rewards increase?

PR #1086, opened 2019-12-23
fix a minor bug in mpow calculation

tigers334:
This fix might lead to a slight efficiency gain in MPoW calculation.

### DIFF
--- a/consensus/spow/engine.go
+++ b/consensus/spow/engine.go
@@ -663,7 +663,7 @@ func calDetmLoop(matrix *mat.Dense, dim int, log *log.SeeleLog) (float64, int) {
 				return det, nonZerosCount
 			}
 			// at this point, even all left are ok, the total is still smaller than target, just stop!
-			if nonZerosCount+(256-i) < nonZeroCountTarget {
+			if nonZerosCount+(256-i-dim) < nonZeroCountTarget {
 				return det, nonZerosCount
 			}
 		}


LincolnZeng:
Thank you very much. @tigers334 
But the for loop will end at (256-dim), so should use ''nonZerosCount+(256-i)" other than "nonZerosCount+(256-i-dim)"
```
	for i := 0; i < 256-dim; i++ {
		submatrix = submatCopy(matrix, i, dim)

		det := mat.Det(submatrix)
		// return first det
		if i == 0 {
			ret = det
		} else {
			// check number of det whose det is larger than 0
			detInt := int64(det)
			detBig := big.NewInt(detInt)
			if detBig.Cmp(big.NewInt(0)) > 0 {

				nonZerosCount++
			}
			// already meet the requirement, just stop and return
			if nonZerosCount >= nonZeroCountTarget {
				return det, nonZerosCount
			}
			// at this point, even all left are ok, the total is still smaller than target, just stop!
			if nonZerosCount+(256-i) < nonZeroCountTarget {
				return det, nonZerosCount
			}
		}
	}
```

tigers334:
Dear LincolnZeng,

Thank you for your comment.
I regard the if statement is for early return for hopeless matrices whose
non-zero count would be surely less than the target.
Is my understanding correct?
if correct, I think it should be "nonZerosCount+(256-i-dim)".
By the way, is the possible maximum non-zero count is (256-dim)=226, right?

Best regards,

2019年12月24日(火) 9:36 LincolnZeng <notifications@github.com>:

tigers334:
Let me explain why it should be "nonZerosCount+(256-i-dim)" using the following example.

Now, we have the variables as below:
i = 200, nonZeroCount = 80, nonZeroCountTarget = 119
Then, in the IF statement of the original code, the left side is
nonZeroCount+(256-i) = 136
This is greater than the right side, which is nonZeroCountTarget (= 119).
Thus, this loop continues.

However, in reality, this loop should be stopped at this point because there is no chance to reach the nonZeroCountTarget in the next loops.
As declared in the FOR statement, the maximum number of iterations of this loop is 226(=256-dim).
This means we have only 26(=226-i) chances to increment nonZeroCount and the possible maximum nonZeroCount is just 106(=80+26).
The number of left loops is obviously hopeless and we can stop the loop.

LincolnZeng:
You are right, once we are at the position of nonZeroCount+(256-dim-i), should not continue to calculate, this will increase the efficiency even though how much it will is not clear, If you can commit a pull request to change it, we can merge into the master after peers reviewing.

PR #1089, opened 2020-01-04
Sync optimization

horusseelle:


light/peer.go
```diff
@@ -197,6 +200,13 @@ func (p *peer) handleDownloadHeadersRequest(msg *DownloadHeaderQuery) error {
 
 	buff := common.SerializePanic(sendMsg)
 	p.log.Debug("peer send [downloadHeadersResponseCode] query with size %d byte", len(buff))
+	if !p.Peer.GetConnStat(){
+        return errors.New("peer not connected")
+	}
+	
+	if !p.Peer.GetConnStat(){
+        return nil
+	}
```
shawnxin89: The second if isn't used.

p2p/peer.go
```diff
@@ -152,6 +154,16 @@ func (p *Peer) readLoop(readErr chan<- error) {
...
+func (p *Peer) notifyProtocolsPeerVerify() {
+	for _, proto := range p.protocolMap {
+		if proto.GetPeer(p.Node.ID) != nil {
+			continue
+		} else {
+
+			proto.AddPeer(p, &proto)
+		}
+	}
+}
```
shawnxin89: Need a lock here?

seele/peer.go
```diff
@@ -120,7 +123,13 @@ func (p *peer) sendTransactionHash(txHash common.Hash) error {
 		return nil
 	}
 	buff := common.SerializePanic(txHash)
+    if !p.Peer.GetConnStat(){
+        return errors.New("peer not connected")
+	}
 
+	if !p.Peer.GetConnStat(){
+        return errors.New("peer not connected")
+	}
```
shawnxin89: The second if isn't used.

PR #1094, opened 2020-01-22
Introduce speculative calculation of determinants

tigers334:
There are a minor change and a major one in the 2 commits.

1. Skipping determinant of the first submatrix
This is namely skipping the calculation of the first submatrix's determinant because the first determinant seems not to contribute MPoW at all.

2. Introduction of speculative determinant calculation
Recently the difficulty is so high that the most of all determinant calculation for just counting non-zero ones becomes in vain.
Thus, I suggest calculation of determinants in the last submatrices in the first place.
If there were no great determinants which surpass the difficulty target, the loop stops.
In this manner, we can cut down on over 90% calculation of determinants.
Please look over the code change.

Could you validate the second commit especially?

### DIFF
--- a/consensus/spow/engine.go
+++ b/consensus/spow/engine.go
@@ -540,7 +540,7 @@ miner:
 				header.Witness = []byte(strconv.FormatUint(nonce-uint64(dim-1), 10))
 				hash = header.Hash()
 			}
-			res, count := calDetmLoop(matrix, dim, log)
+			res, count := calDetmLoopForMining(matrix, dim, target, log)
 			restInt := int64(res)
 			restBig := big.NewInt(restInt)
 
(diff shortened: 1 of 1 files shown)


shawnxin89:
> There are a minor change and a major one in the 2 commits.
> 
> 1. Skipping determinant of the first submatrix
>    This is namely skipping the calculation of the first submatrix's determinant because the first determinant seems not to contribute MPoW at all.
> 2. Introduction of speculative determinant calculation
>    Recently the difficulty is so high that the most of all determinant calculation for just counting non-zero ones becomes in vain.
>    Thus, I suggest calculation of determinants in the last submatrices in the first place.
>    If there were no great determinants which surpass the difficulty target, the loop stops.
>    In this manner, we can cut down on over 90% calculation of determinants.
>    Please look over the code change.
> 
> Could you validate the second commit especially?


Thank you for the commit!

